PR #1, opened 2024-05-04
Add type specialization for performance

efaulhaber:
Previously, we circumvented this by making everything inline.

@sloede @ranocha which is the better approach? Inlining everything so that there is no non-inlined function where Julia avoids specialization, or this PR?


sloede:
Inlining depends on the compiler and is almost always heuristics based. It also won't work if you use debugging mode without Inlining to catch some errors. Thus if you have found that the presently implemented approach does really speed things up, I'd stick with it.

ranocha:
Benchmark and see what's better?

LasNikas (approved):

PR #2, opened 2024-05-04
Remove thread sleeps that were needed for old Julia versions on macOS ARM

efaulhaber:
These made the lagging issues a little better, but slowed down other architectures. For Julia 1.10, this is no longer needed.
I tested Julia 1.9, and the lagging is bad either way, so I don't see any reason to keep this. Discouraging users from using Julia 1.9 on macOS ARM is probably the better solution.


sloede:
I concur with your assessment 👍

PR #6, opened 2024-05-08
Rename package to PointNeighbors.jl

efaulhaber:


sloede (approved):
LGTM! I didn't check everything in detail, but the most important places seem to be covered.

PR #9, opened 2024-05-10
Implement neighborhood search based on static neighbor lists

efaulhaber:
I finally properly implemented the prototype in https://github.com/trixi-framework/TrixiParticles.jl/pull/284.

Based on #13.

### DIFF
--- a/src/nhs_neighbor_lists.jl
+++ b/src/nhs_neighbor_lists.jl
@@ -1,7 +1,7 @@
 @doc raw"""
-    NeighborListsNeighborhoodSearch{NDIMS}(search_radius, n_particles;
-                                           periodic_box_min_corner = nothing,
-                                           periodic_box_max_corner = nothing)
+    PrecomputedNeighborhoodSearch{NDIMS}(search_radius, n_particles;
+                                         periodic_box_min_corner = nothing,
+                                         periodic_box_max_corner = nothing)
 
 Neighborhood search with precomputed neighbor lists. A list of all neighbors is computed
 for each particle during initialization and update.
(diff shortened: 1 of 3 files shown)


src/nhs_neighbor_lists.jl
```diff
@@ -0,0 +1,115 @@
...
+        pos_diff, distance2 = compute_periodic_distance(pos_diff, distance2, search_radius,
+                                                        periodic_box)
+
+        distance = sqrt(distance2)
+
+        # Inline to avoid loss of performance
+        # compared to not using `for_particle_neighbor`.
+        @inline f(particle, neighbor, pos_diff, distance)
+    end
+end
```
LasNikas, line 115: Can't you just add `eachneighbor(particle_coords, particle, neighborhood_search::NeighborListsNeighborhoodSearch) = neighbor_lists[particle]` instead of adding a new method for `foreach_neighbor`?
And maybe add `check_particle_distance(distance, nhs)` but this is not really a performance issue, right?

 You can ignore this suggestion, if it is to cumbersome

src/nhs_neighbor_lists.jl
```diff
@@ -0,0 +1,115 @@
+@doc raw"""
+    NeighborListsNeighborhoodSearch{NDIMS}(search_radius, n_particles;
```
svchb, line 2: Maybe **PrecomputedNeighborhodSearch** would make it easier to understand in the actual application of this method?

svchb (changes requested):

src/nhs_neighbor_lists.jl
efaulhaber: The distance check is performance-relevant. But I might want to add another function like `for_neighbor_index` or so, which skips the whole distance computation stuff (only for this neighborhood search) and only loops over the neighbor index.

src/nhs_neighbor_lists.jl
efaulhaber: Yes, this is good! `NeighborListNeighborhoodSearch` is way too cumbersome. @LasNikas ?

LasNikas: Yes, that is better

LasNikas (approved):

svchb (approved):

PR #13, opened 2024-05-12
Restructure tests

efaulhaber:
The manually constructed periodic tests are now used in two places:
1. For all neighborhood search implementations with `for_particle_neighbor`.
2. For the grid NHS only with `eachneighbor` (only testing for potential neighbors).

PR #15, opened 2024-05-16
Add tests comparing neighbor lists to trivial NHS

efaulhaber:
Based on #13.

### DIFF
--- a/test/point_cloud.jl
+++ b/test/point_cloud.jl
@@ -1,0 +1,30 @@
+using Random
+using StaticArrays
+
+# Generate a rectangular point cloud, optionally with a perturbation in the point positions
+function point_cloud(n_points_per_dimension;
+                     seed = 1, perturbation_factor_position = 1.0)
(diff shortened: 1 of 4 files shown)


test/point_cloud.jl
```diff
@@ -0,0 +1,29 @@
+using Random
+
+# Generate a rectangular point cloud, optionally with a perturbation in the point positions
+function point_cloud(n_points_per_dimension;
+                     seed = 1, perturbation_factor_position = 1.0)
```
LasNikas, line 5: Why is this space between ` = `  not detected by the JuliaFormatter?

efaulhaber: I use a different JuliaFormatter configuration here without `whitespace_in_kwargs = false`. This is consistent with Trixi's new style and SciML, and I thought about changing it for TrixiParticles.jl as well in the future.

LasNikas (approved):
LGTM!

PR #16, opened 2024-05-17
Rename files according to `nhs_*.jl`

LasNikas:
Keeping in mind that more searches will be added, then this could increase readability in the directory, no?


efaulhaber:
I have a local branch depending on #8 and #9 where I move these files to a separate folder `neighborhood_searches`, so that we have `neighborhood_searches/trivial.jl`, `neighborhood_searches/grid.jl` and so on.

efaulhaber (approved):

PR #17, opened 2024-05-17
NHS for triangles or edges

LasNikas:


svchb (approved):

svchb (changes requested):
Please format. Please add test. Also increase the version. Docs are missing.

svchb (changes requested):
Please add test. Also increase the version. Docs are missing.

src/nhs_faces.jl
```diff
@@ -0,0 +1,200 @@
+struct FaceNeighborhoodSearch{NDIMS, ELTYPE, PB}
+    cell_list         :: Dict{NTuple{NDIMS, Int}, Vector{Int}}
```
efaulhaber, line 2: Is there a reason why you cannot use existing cell lists here?

src/nhs_faces.jl
```diff
@@ -0,0 +1,200 @@
...
+        end
+
+        unique!(face_ids)
+
+        if isempty(face_ids)
+            continue
+        end
+
+        neighbor_iterator[cell_runner] = copy(face_ids)
+    end
```
efaulhaber, line 89: Why do you need to explicitly construct these lists compared to just looping over neighboring cells on the fly like the GNHS? Because faces can be in multiple cells, so you need to make it unique somehow?

src/nhs_faces.jl
LasNikas: No, when I started implementing this, PointNeighbors.jl was still living in TrixiParticles.jl and the NHS was not as mature as it is now. I'll switch to `FullGridCellList`

src/nhs_faces.jl
LasNikas: Yes, exactly. Otherwise, I'll loop over the same face multiple times in worse case.

src/nhs_faces.jl
LasNikas: However, I was not sure about `max_points_per_cell`. I mean there can be very large faces among very small faces, so the `max_points_per_cell` may vary very heavily.

LasNikas: also it is very misleading to set max "points" here, as these are actually faces in this case

efaulhaber: Make the cell list a kwarg, which defaults to `DictionaryCellList`, just like the GNHS. Then the default always works, but users can explicitly pass a `FullGridCellList` and manually tweak `max_points_per_cell`.

src/nhs_faces.jl
efaulhaber: Hmmm, then the best idea I have is to use a second cell list to store the neighbors of each cell. This will be more flexible and probably more efficient, but you need to do the `unique!` part manually, I'm afraid, for it to work with `FullGridCellList`.

For a first implementation, you can hardcode a `DictionaryCellList` here.

LasNikas:
After a discussion in slack with @efaulhaber, this now lives in TrixiParticles.jl again.

PR #22, opened 2024-06-04
Add docs for `initialize!` and `update!`

efaulhaber:


src/neighborhood_search.jl
```diff
@@ -1,3 +1,46 @@
+abstract type AbstractNeighborhoodSearch end
+
+"""
+    initialize!(search::AbstractNeighborhoodSearch, x, y)
+
+Initialize a neighborhood search with the two coordinate arrays `x` and `y`.
+
+In general, the purpose of a neighborhood search is to find for one point in `x`
+all points in `y` that are closer to that point than the search radius.
```
LasNikas, line 9:
```suggestion
all points in `y` whose distances to that point are smaller than the search radius.
```
?

src/neighborhood_search.jl
```diff
@@ -1,3 +1,46 @@
...
+
+Not all implementations support incremental updates.
+If incremental updates are not possible for an implementation, `update!` will fall back
+to a regular `initialize!`.
+
+Some neighborhood searches might not need to update when only `x` changed since the last
+`update!` or `initialize!` and `y` did not change. Pass `particles_moving = (true, false)`
+in this case to avoid unnecessary updates.
+The first flag in `particles_moving` indicates if points in `x` are moving.
+The second flag indicates if points in `y` are moving.
```
LasNikas, line 35: This is only the case when using `GridNeighborhoodSearch`, right?
I mean, there are nhs where the search region moves with the particle.

src/neighborhood_search.jl
```diff
@@ -1,3 +1,46 @@
+abstract type AbstractNeighborhoodSearch end
+
+"""
+    initialize!(search::AbstractNeighborhoodSearch, x, y)
+
+Initialize a neighborhood search with the two coordinate arrays `x` and `y`.
+
+In general, the purpose of a neighborhood search is to find for one point in `x`
+all points in `y` that are closer to that point than the search radius.
+`x` and `y` are expected to be matrices, where `x[:, i]` are the coordinates of point `i`.
```
LasNikas, line 10:
```suggestion
`x` and `y` are expected to be matrices, where the columns represent the coordinates of point `i`.
```
?

src/neighborhood_search.jl
efaulhaber: So far:
- `TrivialNeighborhoodSearch`: No update
- `GridNeighborhoodSearch`: Only update when `particles_moving[2]`
- `NeighborListsNeighborhoodSearch`: First update the internally used `GridNeighborhoodSearch` (only when `particles_moving[2]`, then update neighbor lists if `any(particles_moving)`.

LasNikas: Yes, I mean, it might make sense to explicitly mention that this is only the case for nhs where the domain is subdivided into fixed cells, keeping in mind that in future there will be other nhs where update is needed also for static neighbors since the reference point is moving.

efaulhaber: Seems to me like implementation details irrelevant to the user? That's what this option is for. The user just tells the NHS which set of particles are moving and which are static, and they don't have to think about which NHS is used.

LasNikas (approved):
LGTM!

PR #30, opened 2024-06-21
Revise periodic box API

efaulhaber:

PR #32, opened 2024-06-21
Implement `copy_neighborhood_search` to allow empty "template" neighborhood search

### LINKED ISSUE #27: Figure out how to pass neighborhood search to a TrixiParticles.jl simulation
Currently, we only pass the type of the neighborhood search, e.g.:
```julia
semi = Semidiscretization(system1, system2, neighborhood_search=GridNeighborhoodSearch)
```
This comes with two major problems:
1. To use a new neighborhood search in TrixiParticles.jl, one has to add a function `create_neighborhood_search` for this type, which calls the constructor of the neighborhood search and initializes it.
2. Options for the neighborhood search can't be passed. Currently, we only have the `threaded_update` kwarg, which can be passed to `Semidiscretization` and is then passed through `create_neighborhood_search` to the constructor of the NHS. This doesn't work with multiple neighborhood search implementations with different kwargs.

We can't just call the constructor and pass a neighborhood search object because we need multiple objects (one for each pair of physics systems).
My best idea so far is the following:
1. Create and pass to the `Semidiscretization` an empty neighborhood search, setting all options in the constructor. 
2. Then, in PointNeighbors.jl, add a function `copy_neighborhood_search`, which takes a neighborhood search, a number of particles, and a search radius, and then creates a neighborhood search with the same type and options and the new size and search radius.
3. In the NHS constructors, the number of particles is set to zero by default for convenience and the search radius to `nothing` or so. The resulting neighborhood search can't be actually used without a search radius, so it's only to be used with `copy_neighborhood_search`.

The line above would then become
```julia
semi = Semidiscretization(system1, system2, neighborhood_search=GridNeighborhoodSearch())
```
and setting options is easy and modular:
```julia
semi = Semidiscretization(system1, system2, neighborhood_search=GridNeighborhoodSearch(threaded_update=false))
```

For each pair of physics systems, TrixiParticles.jl would then call:
```julia
nhs = copy_neighborhood_search(dummy_nhs, search_radius, n_particles)
```

Does anyone have any other ideas?
@LasNikas @svchb @sloede

efaulhaber:
Based on #30.

Closes #27.


src/nhs_neighbor_lists.jl
```diff
@@ -1,6 +1,6 @@
 @doc raw"""
```
LasNikas, line 1: Should the file name better be `nhs_precomputed.jl`?

LasNikas (changes requested):

LasNikas (approved):

PR #25, opened 2024-06-12
Add "full grid" cell list with a separate list for each cell of a rectangular domain

efaulhaber:
Based on #12.

PR #34, opened 2024-06-24
Implement dynamic vector of vectors data structure

efaulhaber:
This is similar to `VectorOfVectors` of ArraysOfArray.jl, but dynamic, supporting `push!` and `deleteat!`.
This is needed to make #25 GPU-compatible (and more efficient in general).
Coincidentally, the update buffer was using a similar data structure, so we already have a first use-case for this new data structure.

### DIFF
--- a/src/nhs_grid.jl
+++ b/src/nhs_grid.jl
@@ -47,15 +47,14 @@ since not sorting makes our implementation a lot faster (although less paralleli
   In: Computer Graphics Forum 30.1 (2011), pages 99–112.
   [doi: 10.1111/J.1467-8659.2010.01832.X](https://doi.org/10.1111/J.1467-8659.2010.01832.X)
 """
-struct GridNeighborhoodSearch{NDIMS, ELTYPE, CL, PB} <: AbstractNeighborhoodSearch
-    cell_list           :: CL
-    search_radius       :: ELTYPE
-    periodic_box        :: PB
-    n_cells             :: NTuple{NDIMS, Int}    # Required to calculate periodic cell index
-    cell_size           :: NTuple{NDIMS, ELTYPE} # Required to calculate cell index
-    cell_buffer         :: Array{NTuple{NDIMS, Int}, 2} # Multithreaded buffer for `update!`
-    cell_buffer_indices :: Vector{Int} # Store which entries of `cell_buffer` are initialized
-    threaded_update     :: Bool
+struct GridNeighborhoodSearch{NDIMS, ELTYPE, CL, PB, UB} <: AbstractNeighborhoodSearch
+    cell_list       :: CL
+    search_radius   :: ELTYPE
+    periodic_box    :: PB
+    n_cells         :: NTuple{NDIMS, Int}    # Required to calculate periodic cell index
+    cell_size       :: NTuple{NDIMS, ELTYPE} # Required to calculate cell index
+    update_buffer   :: UB # Multithreaded buffer for `update!`
+    threaded_update :: Bool
 
     function GridNeighborhoodSearch{NDIMS}(; search_radius = 0.0, n_points = 0,
                                            periodic_box = nothing,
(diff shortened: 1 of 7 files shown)


src/nhs_grid.jl
```diff
@@ -86,37 +87,28 @@ struct GridNeighborhoodSearch{NDIMS, ELTYPE, CL, PB} <: AbstractNeighborhoodSear
...
                      x::AbstractMatrix, y::AbstractMatrix)
     initialize_grid!(neighborhood_search, y)
 end
 
-function initialize_grid!(neighborhood_search::GridNeighborhoodSearch{NDIMS},
-                          y::AbstractMatrix) where {NDIMS}
-    initialize_grid!(neighborhood_search, i -> extract_svector(y, Val(NDIMS), i))
-end
-
-function initialize_grid!(neighborhood_search::GridNeighborhoodSearch, coords_fun)
```
efaulhaber, line 112: Initialization via a coordinate function was a feature that I kept because "why not?", but now it's in the way, and it's not used anyway, so I removed it.

src/vector_of_vectors.jl
```diff
@@ -0,0 +1,96 @@
+# Data structure that behaves like a `Vector{Vector}`, but uses a contiguous memory layout.
+# Similar to `VectorOfVectors` of ArraysOfArrays.jl, but allows to resize the inner vectors.
+struct DynamicVectorOfVectors{T, ARRAY2D, ARRAY1D} <: AbstractVector{Array{T, 1}}
```
efaulhaber, line 3: I'm open for better naming suggestions.

src/vector_of_vectors.jl
```diff
@@ -0,0 +1,96 @@
...
+    return vov
+end
+
+@inline function Base.push!(vov::DynamicVectorOfVectors, vector::AbstractVector, vectors...)
+    push!(vov, vector)
+    push!(vov, vectors...)
+end
+
+# `push!(vov[i], value)`
+@inline function pushat!(vov::DynamicVectorOfVectors, i, value)
```
efaulhaber, line 52: We could use `push!(vov[i], value)` and `deleteat!(vov[i], j)` by returning a wrapper around a view, which also stores the `vov` and the column where the view comes from. This wrapper would behave like a regular view, but allow `push!` and `deleteat!`. Not sure if that would be more elegant than this here. We only use this data structure internally, anyway.

sloede: > We only use this data structure internally, anyway.

Famous last words.

test/vector_of_vectors.jl
```diff
@@ -0,0 +1,78 @@
...
+    # Test different types by defining a function to convert to this type
+    types = [Int32, Float64, i -> (i, i)]
+
+    @testset verbose=true "Eltype $(eltype(type(1)))" for type in types
+        ELTYPE = typeof(type(1))
+        vov_ref = Vector{Vector{ELTYPE}}()
+        vov = PointNeighbors.DynamicVectorOfVectors{ELTYPE}(max_outer_length = 20,
+                                                            max_inner_length = 10)
+
+        function verify(vov, vov_ref)
```
LasNikas, line 11: Very nice tests!
Could you maybe add `sizeof(vov) == ....` ?
That's the interesting point, no?

src/vector_of_vectors.jl
svchb: NestedDynamicVector?

svchb: ContiguousVectorMatrix

svchb (dismissed):

LasNikas (approved):

src/vector_of_vectors.jl
efaulhaber: I don't think any of these are more descriptive. I have to close the conversation to merge, but I'm still open for suggestions.

PR #36, opened 2024-06-25
Some minor remarks/suggestions on #18

LasNikas:
Some minor remarks/suggestions on #18

### DIFF
--- a/benchmarks/count_neighbors.jl
+++ b/benchmarks/count_neighbors.jl
@@ -11,6 +11,15 @@ Due to the minimal computational cost, differences between neighborhood search
 implementations are highlighted. On the other hand, this is the least realistic benchmark.
 
 For a computationally heavier benchmark, see [`benchmark_n_body`](@ref).
+
+# Example
+```julia
+    nhs_trivial = TrivialNeighborhoodSearch .......
+
+    ...
+
+    nhs_grid = GridNeighborhoodSearch ....
+```
 """
 function benchmark_count_neighbors(neighborhood_search, coordinates; parallel = true)
     n_neighbors = zeros(Int, size(coordinates, 2))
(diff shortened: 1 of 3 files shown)


benchmarks/count_neighbors.jl
```diff
@@ -11,6 +11,15 @@ Due to the minimal computational cost, differences between neighborhood search
...
 For a computationally heavier benchmark, see [`benchmark_n_body`](@ref).
+
+# Example
+```julia
+    nhs_trivial = TrivialNeighborhoodSearch .......
+
+    ...
+
+    nhs_grid = GridNeighborhoodSearch ....
+```
```
efaulhaber, line 22: I didn't add an example because this benchmark is mostly to be used with `plot_benchmarks`.

benchmarks/plot.jl
```diff
@@ -28,10 +28,16 @@ of points and plot the results.
...
+
+# Example
+
+also an example?
 """
 function plot_benchmarks(benchmark, n_points_per_dimension, iterations;
                          parallel = true, title = "",
                          seed = 1, perturbation_factor_position = 1.0)
+    # Is there a reason why this is hard coded? Why not passing custom nhs like `custom_nhs...`?
+    # And then `nameof(typeof(nhs))`
```
efaulhaber, line 40: 1. You would have to pass an array of functions `(search_radius, coordinates) -> nhs`, which would get ugly quickly.
2. `nameof(typeof(nhs))` is not enough. With new backends and options, many of these names will be "`SomeNeighborhoodSearch` with `some_option=true`".

I figured it is easier to treat this like an example file that you modify and include again.

svchb:
@LasNikas Can we close this?

PR #42, opened 2024-06-26
Implement fully parallel update with atomic operations

efaulhaber:
Based on #40.

grafik
grafik
grafik
grafik

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,12 +4,14 @@ authors = ["Erik Faulhaber <erik.faulhaber@uni-koeln.de>"]
 version = "0.3.2-dev"
 
 [deps]
+Atomix = "a9b6321e-bd34-4604-b9c9-b65b8de01458"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Polyester = "f517fe37-dbe3-4b94-8317-1923a5111588"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
+Atomix = "0.1"
 LinearAlgebra = "1"
 Polyester = "0.7.5"
 Reexport = "1"
--- a/src/cell_lists/dictionary.jl
+++ b/src/cell_lists/dictionary.jl
@@ -25,6 +25,8 @@ struct DictionaryCellList{NDIMS} <: AbstractCellList
     end
 end
 
+supported_update_strategies(::DictionaryCellList) = (:semi_parallel, :serial)
+
 function Base.empty!(cell_list::DictionaryCellList)
     Base.empty!(cell_list.hashtable)
 
--- a/src/cell_lists/dictionary.jl
+++ b/src/cell_lists/dictionary.jl
@@ -25,7 +25,7 @@ struct DictionaryCellList{NDIMS} <: AbstractCellList
     end
 end
 
-supported_update_strategies(::DictionaryCellList) = (:semi_parallel, :serial)
+supported_update_strategies(::DictionaryCellList) = (SemiParallelUpdate, SerialUpdate)
 
 function Base.empty!(cell_list::DictionaryCellList)
     Base.empty!(cell_list.hashtable)
--- a/src/nhs_grid.jl
+++ b/src/nhs_grid.jl
@@ -37,19 +37,12 @@ since not sorting makes our implementation a lot faster (although less paralleli
 - `cell_list`:              The cell list that maps a cell index to a list of points inside
                             the cell. By default, a [`DictionaryCellList`](@ref) is used.
 - `update_strategy = nothing`: Strategy to parallelize `update!`. Available options are:
-    - `nothing`:        Automatically choose the best available option.
-    - `:parallel`:      Fully parallel update by using atomic operations to avoid race
-                        conditions when adding points into the same cell. This is not
-                        available for all cell list implementations, but is the default when
-                        available.
-    - `:semi-parallel`: Loop over all cells in parallel to mark cells with points that now
-                        belong to a different cell. Then, move points of affected cells
-                        serially. This is available for all cell list implementations and is
-                        the default when `:parallel` is not available.
-    - `:serial`:        Deactivate parallelization in the neighborhood search update.
-                        Parallel update can be one of the largest sources of variations
-                        between simulations with different thread numbers due to neighbor
-                        ordering changes.
+    - `nothing`: Automatically choose the best available option.
+    - [`ParallelUpdate()`](@ref): This is not available for all cell list implementations,
+        but is the default when available.
+    - [`SemiParallelUpdate()`](@ref): This is available for all cell list implementations
+        and is the default when [`ParallelUpdate`](@ref) is not available.
+    - [`SerialUpdate()`](@ref)
 
 ## References
 - M. Chalela, E. Sillero, L. Pereyra, M.A. Garcia, J.B. Cabral, M. Lares, M. Merchán.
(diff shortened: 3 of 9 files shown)


src/nhs_grid.jl
```diff
@@ -219,6 +253,50 @@ end
     end
 end
 
+# Fully parallel update with atomic push
+function update_grid!(neighborhood_search::GridNeighborhoodSearch{<:Any, :parallel},
+                      coords_fun::Function)
+    (; cell_list) = neighborhood_search
+
+    # Note that we need two separate loops for adding and removing points.
+    # `push_cell_atomic!` only guarantees thread-safety when different threads push at the
```
LasNikas, line 262:
```suggestion
    # `push_cell_atomic!` guarantees thread-safety when different threads push
```

LasNikas (commented):
IMO, it might make sense to export `supported_update_strategies(cell_list)` or add to the docs that the supported strategies depends on the `cell_list` type if I understood correctly.

Project.toml
```diff
@@ -4,12 +4,14 @@ authors = ["Erik Faulhaber <erik.faulhaber@uni-koeln.de>"]
 version = "0.3.2-dev"
 
 [deps]
+Atomix = "a9b6321e-bd34-4604-b9c9-b65b8de01458"
```
svchb, line 7: Are you sure about `Atomix`? Seems to be dead.
You could just copy the reevant macros from sugar.jl https://github.com/JuliaConcurrent/Atomix.jl/blob/main/src/sugar.jl
Also there seem to be issues on GPUs?

efaulhaber: It's not dead, it just didn't have to be modified in the past 2 years because it's a simple package.
KernelAbstractions.jl is also based on it:
https://github.com/JuliaGPU/KernelAbstractions.jl/blob/4285051c3ea949aee8a024383f38c8849f602729/src/KernelAbstractions.jl#L12

src/nhs_grid.jl
```diff
@@ -51,25 +61,44 @@ since not sorting makes our implementation a lot faster (although less paralleli
...
-    cell_list       :: CL
-    search_radius   :: ELTYPE
-    periodic_box    :: PB
-    n_cells         :: NTuple{NDIMS, Int}    # Required to calculate periodic cell index
-    cell_size       :: NTuple{NDIMS, ELTYPE} # Required to calculate cell index
-    update_buffer   :: UB # Multithreaded buffer for `update!`
-    threaded_update :: Bool
+struct GridNeighborhoodSearch{NDIMS, update_strategy,
+                              ELTYPE, CL, PB, UB} <: AbstractNeighborhoodSearch
+    # Store `update_strategy` as type parameter to dispatch `update!`
```
efaulhaber, line 66: I know that @sloede is not a fan of storing things as type parameters. I find it quite elegant that we can dispatch on the update strategy.

src/nhs_grid.jl
efaulhaber: This is correct. It only guarantees thread-safety for parallel push, but not when deletes happen as well.

LasNikas: but "at the" makes no sense to me, semantically.

efaulhaber: Ah, oops. Looks like I couldn't decide between "at the same time" and "simultaneously".

src/cell_lists/dictionary.jl
```diff
@@ -25,6 +25,8 @@ struct DictionaryCellList{NDIMS} <: AbstractCellList
     end
 end
 
+supported_update_strategies(::DictionaryCellList) = (:semi_parallel, :serial)
```
sloede, line 28: Why do you return tuples of symbols instead of distinct values that can be used for type-stable dispatch? E.g., along the lines of
```julia
struct UpdateStrategy{T} end
const semi_parallel = UpdateStrategy{:semi_parallel}()
const serial = UpdateStrategy{:serial}()

...

foo(::UpdateStrategy{:semi_parallel}) = "only semi_parallel"
bar(::Union{UpdateStrategy{:semi_parallel}, UpdateStrategy{:serial}}) = "serial or semi_parallel"

foo(semi_parallel)
bar(serial)
```
?

Of course, still observe YAGNI, so maybe the current approach is sufficient

src/nhs_grid.jl
```diff
@@ -36,10 +36,20 @@ since not sorting makes our implementation a lot faster (although less paralleli
...
-                            neighborhood search update. This can be one of the largest
-                            sources of variations between simulations with different
-                            thread numbers due to neighbor ordering changes.
+- `update_strategy = nothing`: Strategy to parallelize `update!`. Available options are:
+    - `nothing`:        Automatically choose the best available option.
+    - `:parallel`:      Fully parallel update by using atomic operations to avoid race
+                        conditions when adding points into the same cell. This is not
+                        available for all cell list implementations, but is the default when
+                        available.
+    - `:semi-parallel`: Loop over all cells in parallel to mark cells with points that now
```
sloede, line 45:
```suggestion
    - `:semi_parallel`: Loop over all cells in parallel to mark cells with points that now
```

src/nhs_grid.jl
```diff
@@ -36,10 +36,20 @@ since not sorting makes our implementation a lot faster (although less paralleli
...
+                        belong to a different cell. Then, move points of affected cells
+                        serially. This is available for all cell list implementations and is
+                        the default when `:parallel` is not available.
+    - `:serial`:        Deactivate parallelization in the neighborhood search update.
+                        Parallel update can be one of the largest sources of variations
+                        between simulations with different thread numbers due to neighbor
+                        ordering changes.
```
sloede, line 52:
```suggestion
    - `:serial`:        Deactivate parallelization in the neighborhood search update.
                        Parallel update can be one of the largest sources of runtime variations
                        between simulations with different thread numbers due to neighbor
                        ordering changes.
```

src/nhs_grid.jl
sloede: I am not against storing things as type parameters if they can be used for dispatch. It's just that in my mind, `snake_case` names are strongly associated with variable and function names, which is why I find it somewhat weird. But I think it makes sense here.

src/nhs_grid.jl
efaulhaber: That's annoying to work with:
```
julia> :semi-parallel
ERROR: UndefVarError: `parallel` not defined
Stacktrace:
 [1] top-level scope
   @ REPL[9]:1
```

src/cell_lists/dictionary.jl
efaulhaber: How about
```julia
struct ParallelUpdate end
struct SemiParallelUpdate end
struct SerialUpdate

struct NeighborhoodSearch{U}
    update_strategy :: U
end

supported_update_strategies() = (ParallelUpdate(), SemiParallelUpdate(), SerialUpdate())

update!(nhs) = update!(nhs, nhs.update_strategy)
update!(nhs, ::ParallelUpdate) = ...
update!(nhs, ::Union{SerialUpdate, SemiParallelUpdate}) = ...
```
and then
```julia
nhs = NeighborhoodSearch(update_strategy = SerialUpdate())
```

sloede: Yes, much better!

src/nhs_grid.jl
efaulhaber: "one of the largest sources of *error variations* between simulations with different thread numbers"?

sloede: Ah, then I misread it. Either way, it was not clear before what kind of variations we're talking about here, thus it would be great if you can clarify it

LasNikas (approved):

PR #45, opened 2024-06-27
Make `GridNeighborhoodSearch` with `FullGridCellList` run on the GPU

efaulhaber:
Based on #43.

This PR will finally make the `GridNeighborhoodSearch` with the `FullGridCellList` and the `DynamicVectorOfVectors` backend run on the GPU.

### DIFF
--- a/src/neighborhood_search.jl
+++ b/src/neighborhood_search.jl
@@ -131,18 +131,44 @@ See also [`initialize!`](@ref), [`update!`](@ref).
 """
 function foreach_point_neighbor(f::T, system_coords, neighbor_coords, neighborhood_search;
                                 points = axes(system_coords, 2),
-                                parallel = true) where {T}
+                                parallel::Union{Bool, KernelAbstractions.Backend} = true) where {
+                                                                                                 T
+                                                                                                 }
     # The type annotation above is to make Julia specialize on the type of the function.
     # Otherwise, unspecialized code will cause a lot of allocations
     # and heavily impact performance.
     # See https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing
+    if parallel isa Bool
+        # When `false` is passed, run serially. When `true` is passed, run either a
+        # threaded loop with `Polyester.@batch`, or, when `system_coords` is a GPU array,
+        # launch the loop as a kernel on the GPU.
+        parallel_ = Val(parallel)
+    elseif parallel isa KernelAbstractions.Backend
+        # WARNING! Undocumented, experimental feature:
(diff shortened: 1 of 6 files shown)


src/util.jl
```diff
@@ -23,53 +23,74 @@ end
...
+        @inline f(iterator[indices[i]])
+    end
+
+    KernelAbstractions.synchronize(backend)
+end
+
+@kernel function generic_kernel(f)
+    i = @index(Global)
+    @inline f(i)
+end
```
efaulhaber, line 96: I was thinking about making this a package extension, so that we don't add KernelAbstractions.jl as a dependency. However, then users would get unspecific "scalar indexing" errors when they forget to load KernelAbstractions.jl.

This function dispatches on `KernelAbstractions.Backend`, so without importing KernelAbstractions, I can't do this:
```julia
@inline function parallel_foreach(f, iterator,
                                  x::Union{AbstractGPUArray, KernelAbstractions.Backend})
    error("Load KernelAbstractions.jl")
end
```

Thoughts? @sloede ?

src/neighborhood_search.jl
```diff
@@ -131,18 +131,44 @@ See also [`initialize!`](@ref), [`update!`](@ref).
 """
 function foreach_point_neighbor(f::T, system_coords, neighbor_coords, neighborhood_search;
                                 points = axes(system_coords, 2),
-                                parallel = true) where {T}
+                                parallel::Union{Bool, KernelAbstractions.Backend} = true) where {
+                                                                                                 T
+                                                                                                 }
```
svchb, line 136: {
T
}
Looks weird

src/util.jl
```diff
@@ -23,53 +23,74 @@ end
...
+macro threaded(system, expr)
+    # Reverse-engineer the for loop.
+    # `expr.args[1]` is the head of the for loop, like `i = eachindex(x)`.
+    # So, `expr.args[1].args[2]` is the iterator `eachindex(x)`
+    # and `expr.args[1].args[1]` is the loop variable `i`.
+    iterator = expr.args[1].args[2]
+    i = expr.args[1].args[1]
+    inner_loop = expr.args[2]
 
+    # Assemble the for loop again as a call to `parallel_foreach`, using `$i` to use the
```
svchb, line 54:
```suggestion
    # Assemble the `for` loop again as a call to `parallel_foreach`, using `$i` to use the
```

svchb (changes requested):

src/neighborhood_search.jl
efaulhaber: I can switch the argument order.

svchb (approved):

LasNikas (approved):

PR #55, opened 2024-07-09
Add content to README

efaulhaber:


sloede (approved):
LGTM! I recommend to add an authors section for easily finding the right person to contact in case of question.

LasNikas (approved):

PR #61, opened 2024-07-15
Additional backend types to support Base.Threads

kaipartmann:
As discussed last week at JuliaCon, specifying the multithreading backend could be very beneficial for users. With the current approach of using Polyester in the `@threaded` macro, a user is forced to either use Polyester or only a serial update. The performance impact of mixing different multithreading backends is also described in https://github.com/kaipartmann/Peridynamics.jl/issues/110.

This PR adds additional types that could be specified as parallelization backend. I tested the code with `Peridynamics.jl` and noticed a slight performance improvement using `Threads.@threads :static` with 64 Threads compared to `Polyester.@batch`.

I did not add unit tests, as I was unsure where to include them correctly in your current testing setup. If you point me in the right direction, I will also include them.


svchb:
Is the performance increase a general improvement or only in the mixed Polyester and Threads case?

kaipartmann:
@svchb I used the following Benchmark with 64 Threads:

```julia
using Peridynamics

function sphere_impact(; ns=6, np=3, path)
    Ø = 0.15
    ΔX0_sphere = Ø / ns
    pos_sphere, vol_sphere = uniform_sphere(Ø, ΔX0_sphere; center_z=Ø / 2 + ΔX0_sphere)
    sphere = Body(BBMaterial{EnergySurfaceCorrection}(), pos_sphere, vol_sphere)
    failure_permit!(sphere, false)
    material!(sphere; horizon=3.1ΔX0_sphere, E=210e9, rho=8000, Gc=2000)
    velocity_ic!(sphere, :all_points, :z, -20)
    lxy, lz = 2.0, 0.1
    ΔX0_plate = lz / np
    pos_plate, vol_plate = uniform_box(lxy, lxy, lz, ΔX0_plate; center_z=-lz / 2)
    plate = Body(BBMaterial{EnergySurfaceCorrection}(), pos_plate, vol_plate)
    material!(plate; horizon=3.1ΔX0_plate, E=27e9, rho=2700, Gc=10)
    ms = MultibodySetup(:sphere => sphere, :plate => plate)
    contact!(ms, :sphere, :plate; radius=min(ΔX0_sphere, ΔX0_plate))
    vv = VelocityVerlet(steps=2000)
    job = Job(ms, vv; path=path)
    @time submit(job)
    return nothing
end

##
path = "results/benchmarks/sphere_impact"
rm(path; recursive=true, force=true)
sphere_impact(; path) # compilation
sphere_impact(; ns=20, np=8, path) # work
```

Due to the Polyester-Threads mixing issue I thought it could be a good idea to also use `Polyester.@batch`, so it is used for every multithreading loop in the package in `v0.3.0`:
```julia
## Peridynamics.jl v0.3.0 - everywhere using Polyester.@batch
# compilation:
# 89.256836 seconds (47.14 M allocations: 14.914 GiB, 0.66% gc time, 7.14% compilation time)
# work:
# 45.514108 seconds (40.16 M allocations: 128.786 GiB, 8.94% gc time)
```

However, I found out this was the reason my parallel performance for single body simulations is not as good as before. When changing all `Polyester.@batch` statements in `Peridynamics.jl` to `@threads :static` calls, the performance of contact simulations also improves (if `PointNeighbors.jl` also uses `@threads :static`):
```julia
## Peridynamics.jl v0.3.1-DEV - everywhere using Threads.@threads :static
## PointNeighbors.jl v0.5.5-dev: update_grid!(...; parallelization_backend=ThreadsStaticBackend())
# compilation:
# 106.717161 seconds (54.34 M allocations: 15.806 GiB, 0.61% gc time, 5.13% compilation time)
# work:
# 39.184317 seconds (48.59 M allocations: 129.752 GiB, 9.53% gc time)
```

@efaulhaber, I also tried specifying `KernelAbstractions.CPU(static=false)` as you wrote me on slack, which is just slightly slower than `@threads :static` and the compilation takes longer:
```julia
## Peridynamics.jl v0.3.1-DEV - everywhere using Threads.@threads :static
## PointNeighbors.jl v0.5.5-dev: update_grid!(...; parallelization_backend=KernelAbstractions.CPU(static=false))
# compilation:
# 122.136295 seconds (52.11 M allocations: 15.556 GiB, 0.55% gc time, 4.67% compilation time)
# work (2 work runs necessary to bench compiled version)
# 40.164468 seconds (47.48 M allocations: 129.602 GiB, 8.91% gc time, 11.92% compilation time)
# 39.528993 seconds (47.38 M allocations: 129.596 GiB, 8.36% gc time)
```

src/neighborhood_search.jl
```diff
@@ -36,6 +36,16 @@ in this case to avoid unnecessary updates.
 The first flag in `points_moving` indicates if points in `x` are moving.
 The second flag indicates if points in `y` are moving.
 
+!!! warning "Experimental feature: Backend specification"
```
efaulhaber, line 39:
```suggestion
!!! warning "Experimental Feature: Backend Specification"
```

src/neighborhood_search.jl
```diff
@@ -36,6 +36,16 @@ in this case to avoid unnecessary updates.
...
+    The keyword argument `parallelization_backend` allows users to specify the
+    multithreading backend. This feature is currently considered experimental!
+
+    Possible parallelization backends are:
+    - [`ThreadsDynamicBackend`](@ref) to use `Threads.@threads :dynamic`
+    - [`ThreadsStaticBackend`](@ref) to use `Threads.@threads :static`
+    - [`PolyesterBackend`](@ref) to use `Polyester.@batch`
+    - `KernelAbstractions.Backend` to execute the update inside a GPU kernel
```
efaulhaber, line 47:
```suggestion
    - `KernelAbstractions.Backend` to launch a GPU kernel
```
Then this is generic and we can copy it to other places later.

src/neighborhood_search.jl
```diff
@@ -141,7 +151,7 @@ function foreach_point_neighbor(f::T, system_coords, neighbor_coords, neighborho
         # threaded loop with `Polyester.@batch`, or, when `system_coords` is a GPU array,
         # launch the loop as a kernel on the GPU.
         parallel_ = Val(parallel)
-    elseif parallel isa KernelAbstractions.Backend
+    elseif parallel isa ParallelizationBackend
         # WARNING! Undocumented, experimental feature:
```
efaulhaber, line 155:
```suggestion
```
Not undocumented anymore.

src/util.jl
```diff
@@ -30,15 +62,21 @@ Semantically the same as `Threads.@threads` when iterating over a `AbstractUnitR
...
+- [`PolyesterBackend`](@ref) to use `Polyester.@batch`
+- [`ThreadsDynamicBackend`](@ref) to use `Threads.@threads :dynamic`
+- [`ThreadsStaticBackend`](@ref) to use `Threads.@threads :static`
+- `KernelAbstractions.Backend` to execute the loop inside a GPU kernel
+
 In particular, the underlying threading capabilities might be provided by other packages
 such as [Polyester.jl](https://github.com/JuliaSIMD/Polyester.jl).
 
-!!! warn
+!!! warning "Not suited for general loops"
```
efaulhaber, line 79:
```suggestion
!!! warning "Warning"
```

src/util.jl
```diff
@@ -67,6 +105,20 @@ end
     end
```
efaulhaber, line 105: I can't comment on the line above, but can you please change the comment to
```
# Use `Polyester.@batch` for low-overhead threading.
# This is currently the default when `x::Array`.

**Review by efaulhaber — CHANGES_REQUESTED** [review]
Thanks a lot for this PR ;)

File: src/util.jl [hunk 6]
```diff
@@ -22,6 +22,38 @@ end
     return floor(Int, i)
 end
 
+abstract type AbstractUserBackend end
```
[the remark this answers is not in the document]
kaipartmann: Renamed to `AbstractThreadingBackend`.

src/util.jl
```diff
@@ -22,16 +22,15 @@ end
...
+abstract type AbstractThreadingBackend end
 
 """
     PolyesterBackend()
 
-A type to specify the parallelization backend when using [`@threaded`](@ref). If this type
-is specified, the @threaded macro will use `Polyester.@batch` for the multithreaded `for`
-loop.
+Pass as first argument to the [`@threaded`](@ref) macro to make the loop multithreaded
+with `Polyester.@batch`.
```
efaulhaber, line 31: Can you please change the other two docstrings as well?

efaulhaber (changes requested):

src/util.jl
kaipartmann: Sorry, I forgot to change them - fixed it.

efaulhaber (approved):
Thanks!

PR #74, opened 2024-11-08
Avoid bounds checking where it is safe to do so

efaulhaber:
This PR and https://github.com/trixi-framework/TrixiParticles.jl/pull/656 yield a ~3% speedup on a Threadripper 3990X and a ~42% speedup on an Nvidia H100. Interestingly, there is no difference on an Nvidia RTX 3090 with FP64 (probably limited by something else), but it benefits even more from inbounds on FP32 with a 60% speedup.

grafik

Not all bounds checks could be safely removed. See #79 for a completely exception-free kernel. The additional speedup is about 2% and can be neglected.

### DIFF
--- a/src/cell_lists/full_grid.jl
+++ b/src/cell_lists/full_grid.jl
@@ -156,7 +156,7 @@ function each_cell_index(cell_list::FullGridCellList{Nothing})
     error("`search_radius` is not defined for this cell list")
 end
 
-@inline function cell_index(cell_list::FullGridCellList, cell::Tuple)
+@propagate_inbounds function cell_index(cell_list::FullGridCellList, cell::Tuple)
     (; linear_indices) = cell_list
 
     return linear_indices[cell...]
@@ -164,7 +164,7 @@ end
 
 @inline cell_index(::FullGridCellList, cell::Integer) = cell
 
-@inline function Base.getindex(cell_list::FullGridCellList, cell)
+@propagate_inbounds function Base.getindex(cell_list::FullGridCellList, cell)
     (; cells) = cell_list
 
     return cells[cell_index(cell_list, cell)]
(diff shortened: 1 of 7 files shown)


src/nhs_grid.jl
```diff
@@ -337,18 +337,37 @@ function update_grid!(neighborhood_search::GridNeighborhoodSearch{<:Any, Paralle
...
 
-    point_coords = extract_svector(system_coords, Val(ndims(neighborhood_search)), point)
     cell = cell_coords(point_coords, neighborhood_search)
 
     for neighbor_cell_ in neighboring_cells(cell, neighborhood_search)
         neighbor_cell = Tuple(neighbor_cell_)
+        neighbors = points_in_cell(neighbor_cell, neighborhood_search)
+
+        for neighbor_ in eachindex(neighbors)
+            neighbor = @inbounds neighbors[neighbor_]
```
svchb, line 367: can this be at the beginning of the line?

src/cell_lists/full_grid.jl
```diff
@@ -156,15 +156,15 @@ function each_cell_index(cell_list::FullGridCellList{Nothing})
...
+@propagate_inbounds function cell_index(cell_list::FullGridCellList, cell::Tuple)
     (; linear_indices) = cell_list
 
     return linear_indices[cell...]
 end
 
 @inline cell_index(::FullGridCellList, cell::Integer) = cell
 
-@inline function Base.getindex(cell_list::FullGridCellList, cell)
+@propagate_inbounds function Base.getindex(cell_list::FullGridCellList, cell)
```
LasNikas, line 167: Is it intended to remove `@inline`? In `util.jl` with `@inbounds`, you kept the `@inline` macro.

efaulhaber: `@propagate_inbounds` implicates `@inline`. In `util.jl` I call a function, which is different.

LasNikas (approved):

PR #76, opened 2024-11-12
Make `max_points_per_cell` work with templates

efaulhaber:


src/cell_lists/full_grid.jl
```diff
@@ -48,12 +48,12 @@ function FullGridCellList(; min_corner, max_corner, search_radius = 0.0,
     # Pad domain to avoid 0 in cell indices due to rounding errors.
     # We can't just use `eps()`, as one might use lower precision types.
     # This padding is safe, and will give us one more layer of cells in the worst case.
-    min_corner = SVector(Tuple(min_corner .- 1e-3 * search_radius))
-    max_corner = SVector(Tuple(max_corner .+ 1e-3 * search_radius))
+    min_corner = SVector(Tuple(min_corner .- 1.0f-3 * search_radius))
+    max_corner = SVector(Tuple(max_corner .+ 1.0f-3 * search_radius))
```
efaulhaber, line 52: This is the only change that is needed to make everything work with `Float32`, so I just included it here.
It does not change anything when `search_radius` is `Float64`.

svchb (approved):

LasNikas (approved):

PR #86, opened 2024-12-04
Use fully parallel atomic initialization

efaulhaber:
Moving away from serial initialization with `FullGridCellList`. I found that this is actually faster for medium to large problems than the incremental update, and it's also much simpler, so I made it the default and added a separate update mode for incremental update.
grafik
I'll have to investigate this in actual simulations and benchmark GPUs before continuing here.


sloede:
Nice speedup!

efaulhaber:
I decided to leave the default at `ParallelIncrementalUpdate` (for now). So this PR does not directly change performance, it only adds a new optional feature.

LasNikas (approved):

svchb (approved):

PR #90, opened 2024-12-23
Add automatic padding for the `FullGridCellList`

### LINKED ISSUE #65: `FullGridCellList` can't handle `NaN` values and throws unspecific errors
See https://github.com/trixi-framework/PointNeighbors.jl/issues/64.
```julia
julia> nhs = GridNeighborhoodSearch{2}(search_radius = 1.0, n_points = 2, cell_list = FullGridCellList(min_corner = (-10.0, -10.0), max_corner = (10.0, 10.0), search_radius = 1.0));

julia> initialize!(nhs, x, x);
ERROR: BoundsError: attempt to access 22×22 LinearIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}} at index [13, -9223372036854775808]
Stacktrace:
 [1] throw_boundserror(A::LinearIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}, I::Tuple{Int64, Int64})
   @ Base ./abstractarray.jl:737
 [2] checkbounds
   @ ./abstractarray.jl:702 [inlined]
 [3] _getindex
   @ ./abstractarray.jl:1323 [inlined]
 [4] getindex
   @ ./abstractarray.jl:1291 [inlined]
 [5] cell_index
   @ /mnt/ssd/home/efaulhab/git/PointNeighbors.jl/src/cell_lists/full_grid.jl:162 [inlined]
 [6] push_cell!
   @ /mnt/ssd/home/efaulhab/git/PointNeighbors.jl/src/cell_lists/full_grid.jl:124 [inlined]
 [7] initialize_grid!(neighborhood_search::GridNeighborhoodSearch{…}, y::Matrix{…})
   @ PointNeighbors /mnt/ssd/home/efaulhab/git/PointNeighbors.jl/src/nhs_grid.jl:176
 [8] initialize!(neighborhood_search::GridNeighborhoodSearch{…}, x::Matrix{…}, y::Matrix{…})
   @ PointNeighbors /mnt/ssd/home/efaulhab/git/PointNeighbors.jl/src/nhs_grid.jl:162
 [9] top-level scope
   @ REPL[14]:1
Some type information was truncated. Use `show(err)` to see complete types.
```
We want to either make this cell list work with `NaN` values or at least throw an error message that properly explains the problem.

The error message when points are outside the bounding box should also be more specific.

### LINKED ISSUE #80: Automatically add padding to `FullGridCellList` bounding box?
When a particle is in a boundary cell of the grid, it is not surrounded by cells, so we get an out-of-bounds exception when looking for neighbor cells. Currently, one has to manually add a padding of `search_radius` to the bounding box.
I think it would be better if the padding was automatically included (only when not using periodicity?) and maybe could be deactivated with `padding = false`.

efaulhaber:
Also adds error messages and closes #65, #80.

### DIFF
--- a/src/cell_lists/full_grid.jl
+++ b/src/cell_lists/full_grid.jl
@@ -50,8 +50,9 @@ function FullGridCellList(; min_corner, max_corner,
     # Also pad domain a little more to avoid 0 in cell indices due to rounding errors.
     # We can't just use `eps()`, as one might use lower precision types.
     # This padding is safe, and will give us one more layer of cells in the worst case.
-    min_corner = SVector(Tuple(min_corner .- 1.001f0 * search_radius))
-    max_corner = SVector(Tuple(max_corner .+ 1.001f0 * search_radius))
+    # `1001 // 1000` is 1.001 without forcing a float type.
+    min_corner = SVector(Tuple(min_corner .- 1001 // 1000 * search_radius))
+    max_corner = SVector(Tuple(max_corner .+ 1001 // 1000 * search_radius))
 
     if search_radius < eps()
         # Create an empty "template" cell list to be used with `copy_cell_list`
--- a/src/cell_lists/full_grid.jl
+++ b/src/cell_lists/full_grid.jl
@@ -45,11 +45,12 @@ supported_update_strategies(::FullGridCellList) = (SemiParallelUpdate, SerialUpd
 function FullGridCellList(; min_corner, max_corner, search_radius = 0.0,
                           backend = DynamicVectorOfVectors{Int32},
                           max_points_per_cell = 100)
-    # Pad domain to avoid 0 in cell indices due to rounding errors.
+    # Add one layer in each direction to make sure neighbor cells exist.
+    # Also pad domain a little more to avoid 0 in cell indices due to rounding errors.
     # We can't just use `eps()`, as one might use lower precision types.
     # This padding is safe, and will give us one more layer of cells in the worst case.
-    min_corner = SVector(Tuple(min_corner .- 1.0f-3 * search_radius))
-    max_corner = SVector(Tuple(max_corner .+ 1.0f-3 * search_radius))
+    min_corner = SVector(Tuple(min_corner .- 1.001f0 * search_radius))
+    max_corner = SVector(Tuple(max_corner .+ 1.001f0 * search_radius))
 
     if search_radius < eps()
         # Create an empty "template" cell list to be used with `copy_cell_list`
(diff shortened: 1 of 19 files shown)


src/nhs_grid.jl
```diff
@@ -203,10 +206,16 @@ end
 # Update only with neighbor coordinates
 function update_grid!(neighborhood_search::GridNeighborhoodSearch{NDIMS},
                       y::AbstractMatrix; parallelization_backend = y) where {NDIMS}
+    check_domain_bounds(neighborhood_search.cell_list, y,
+                        search_radius(neighborhood_search))
```
LasNikas, line 210: In terms of performance, does it make sense to switch off this check for the update?

efaulhaber: You're right. It's slow.

Two options:
1. Make this parallel like so:
```julia
function check_domain_bounds(cell_list::FullGridCellList, y, search_radius)
    valid_ = [true]
    backend = KernelAbstractions.get_backend(y)
    valid = Adapt.adapt(backend, valid_)

    @threaded y for point in 1:size(y, 2)
        coords = extract_svector(y, Val(3), point)
        if any(isnan, coords)
            valid[1] = false
        end

        if any(coords .- search_radius .< cell_list.min_corner) || any(coords .+ search_radius .> cell_list.max_corner)
            valid[1] = false
        end
    end
    if !Vector(valid)[1]
        error("particle coordinates are outside the domain bounds of the cell list")
    end
end
```
The alternating update benchmark yielded:
7.136 ms with Polyester.jl without this check
8.497 ms with Polyester.jl with the check
20.356 ms with CUDA.jl without the check
27.648 ms with CUDA.jl with the check

2. Add a check inside the update loop/kernel. This is what is currently implemented in this PR. This approach does not have a measurable overhead, but has the disadvantage that errors are thrown inside a multithreaded loop. With Polyester.jl, threads are just stopped on errors, but no error message is displayed:
```julia
julia> x = zeros(Int, 5, 10)
5×10 Matrix{Int64}:
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0

julia> @batch for j in axes(x, 2)
           for i in 1:5
               if i > 3 && j in (2, 4, 9)
                   error("some error")
               end
               x[i, j] = 1
           end
       end

julia> x
5×10 Matrix{Int64}:
 1  1  1  1  1  1  1  1  1  1
 1  1  1  1  1  1  1  1  1  1
 1  1  1  1  1  1  1  1  1  1
 1  0  1  0  1  1  1  1  0  1
 1  0  1  0  1  1  1  1  0  1
```
So in this case, particles outside the domain bounds will cause unpredictable behavior.

svchb (changes requested):

src/cell_lists/full_grid.jl
```diff
@@ -42,24 +42,23 @@ end
...
                           max_points_per_cell = 100)
-    # Pad domain to avoid 0 in cell indices due to rounding errors.
+    # Add one layer in each direction to make sure neighbor cells exist.
+    # Also pad domain a little more to avoid 0 in cell indices due to rounding errors.
     # We can't just use `eps()`, as one might use lower precision types.
     # This padding is safe, and will give us one more layer of cells in the worst case.
-    min_corner = SVector(Tuple(min_corner .- 1.0f-3 * search_radius))
-    max_corner = SVector(Tuple(max_corner .+ 1.0f-3 * search_radius))
+    min_corner = SVector(Tuple(min_corner .- 1.001f0 * search_radius))
+    max_corner = SVector(Tuple(max_corner .+ 1.001f0 * search_radius))
```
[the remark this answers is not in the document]
efaulhaber: I changed it and added a comment.

svchb (dismissed):

svchb (approved):

LasNikas (approved):
LGTM!

PR #93, opened 2025-01-16
Drop support for Julia 1.9

efaulhaber:


.github/workflows/ci.yml
```diff
@@ -34,27 +34,19 @@ concurrency:
 
 jobs:
   build:
-    name: Run Tests (Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }})
-    runs-on: ubuntu-latest
+    name: Run Tests (Julia ${{ matrix.version }} - ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
```
efaulhaber, line 38: This was wrong before. All tests were run on Ubuntu (oops).

.github/workflows/ci.yml
```diff
@@ -34,27 +34,19 @@ concurrency:
...
     strategy:
       # Don't cancel all running jobs when one job fails
       fail-fast: false
       matrix:
         version:
-          - '1.9'
-          - '1'
+          - 'min' # the earliest supported version compatible with the project
+          - '1'   # the latest stable 1.x release
```
efaulhaber, line 45: I don't see any reason to run tests for anything between the earliest supported version and latest release. This eliminates the need to change this file every time we increase the minimum version and therefore eliminates the risk of forgetting exactly this.

svchb (approved):

svchb:
You might want to increase version number of the package here as well.

efaulhaber:
I did?

svchb:
Ah you are right I missed it.

LasNikas (approved):

PR #101, opened 2025-04-03
Spatial Hashing

RubberLanding:
Implement spatial hashing data structure for neighborhood search.

Depends on #86.


RubberLanding:
@efaulhaber Please take a look at the current state of the spatial hashing implementation.

RubberLanding:
The problem is that the current code does not set the collision flag correctly for empty cells without any points. 
We only set the collision flag when we call `initialize!()` which in turn calls `push_cell!()` for all points. If there is an empty cell `A` that has a hash collision with a non-empty cell `B`, the collision flag is not set for this hash collision since we did not add a point from `A` where we would notice the collision. 
```julia 
function push_cell!(cell_list::SpatialHashingCellList, cell)
	(; cell_coords, cell_collision, list_size, NDIMS) = cell_list
	key = spatial_hash(cell, list_size)
	cell_coord = cell_coords[key]
	if cell_coord == ntuple(_ -> typemin(Int), NDIMS)
		cell_coords[key] = cell
	elseif cell_coord != cell
		cell_collision[key] = true
	end
end
```

The issue occurs when using `_foreach_neighbor()`, where we iterate over all neighboring cells, and all points in these cells. The points to iterate over are obtained by computing the hash of a neighboring cell `X` and collecting the points stored at this hash location (via `points_in_cell(index)` -> `cell_list[index]`). If the collision flag at this hash location is set, we compute the cell for each point and check if its the same as the current cell `X`. 
If we iterate over an empty cell `A`, we collect all points stored at `A`'s hash location (i.e. since `A` and `B` have a hash collision, all the points that are actually in cell `B`). But since the collision flag is not set, we do not check whether a point actually is in `A`. 

```julia
function check_collision(neighbor_cell_::Tuple, neighbor_coords,
                         cell_list::SpatialHashingCellList, nhs)
    (; list_size, cell_collision) = cell_list

    if cell_collision[spatial_hash(neighbor_cell_, list_size)]
        # Check if `neighbor_coords` are in the cell `neighbor_cell_`
        cell_coords_ = cell_coords(neighbor_coords, nhs)
        return neighbor_cell_ != cell_coords_
    end
    return false
end

@inline function __foreach_neighbor(f, system_coords, neighbor_system_coords,
                                    neighborhood_search::GridNeighborhoodSearch,
                                    point, point_coords, search_radius)
    (; cell_list, periodic_box) = neighborhood_search
    cell = cell_coords(point_coords, neighborhood_search)

    for neighbor_cell_ in neighboring_cells(cell, neighborhood_search)
        neighbor_cell = Tuple(neighbor_cell_)
        neighbors = points_in_cell(neighbor_cell, neighborhood_search)

        for neighbor_ in eachindex(neighbors)
            neighbor = @inbounds neighbors[neighbor_]

            # Making the following `@inbounds` yields a ~2% speedup on an NVIDIA H100.
            # But we don't know if `neighbor` (extracted from the cell list) is in bounds.
            neighbor_coords = extract_svector(neighbor_system_coords,
                                              Val(ndims(neighborhood_search)), neighbor)

            pos_diff = point_coords - neighbor_coords
            distance2 = dot(pos_diff, pos_diff)

            pos_diff, distance2 = compute_periodic_distance(pos_diff, distance2,
                                                            search_radius,
                                                            periodic_box)

            if distance2 <= search_radius^2
                distance = sqrt(distance2)

                # Check if `neighbor_coords` are in the cell `neighbor_cell_`.
                # For the `SpatialHashingCellList`, this might not be the case
                # if we have a collision.
                if check_collision(neighbor_cell_, neighbor_coords, cell_list,
                                   neighborhood_search)
                    continue
                end

                # Inline to avoid loss of performance
                # compared to not using `foreach_point_neighbor`.
                @inline f(point, neighbor, pos_diff, distance)
            end
        end
    end
end
```

efaulhaber:
Would this work?
```julia
function check_collision(neighbor_cell::Tuple, neighbor_coords,
                         cell_list::SpatialHashingCellList, nhs)
    (; list_size, cell_collision) = cell_list

    hash = spatial_hash(neighbor_cell_, list_size)
    if cell_collision[hash]
        # Points from multiple cells are in this list.
        # Check if `neighbor_coords` are in the cell `neighbor_cell`.
        return neighbor_cell != cell_coords(neighbor_coords, nhs)
    elseif cell_coords[hash] != neighbor_cell
        # `cell_collision` is false for this list, meaning only points from one cell are in the list.
        # However, the cell of this list is not our `neighbor_cell`, so `neighbor_coords`
        # are not in `neighbor_cell`.
        return true
    end
    return false
end
```

efaulhaber:
@RubberLanding #86 is merged now. You can go ahead and prepare this PR for review.

src/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,116 @@
...
+    NDIMS::Int
+    points::CL
+    coords::CI
+    collisions::CF
+end
+
+@inline index_type(::SpatialHashingCellList) = Int32
+
+function supported_update_strategies(::PointNeighbors.SpatialHashingCellList)
+    return (SerialUpdate, SemiParallelUpdate)
```
efaulhaber, line 25:
```suggestion
    return (SerialUpdate)
```
Then you say "this NHS only supports reinitialization", and you don't need the special `update!` anymore.

src/nhs_grid.jl
```diff
@@ -210,6 +210,18 @@ function update_grid!(neighborhood_search::GridNeighborhoodSearch{NDIMS},
...
+                                                             <:SpatialHashingCellList},
+                 x::AbstractMatrix,
+                 y::AbstractMatrix; points_moving = (true, true),
+                 parallelization_backend = x)
+    # The coordinates of the first set of points are irrelevant for this NHS.
+    # Only update when the second set is moving.
+    points_moving[2] || return neighborhood_search
+    initialize_grid!(neighborhood_search, y)
+end
+
```
efaulhaber, line 224:
```suggestion
```

efaulhaber (changes requested):

src/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,111 @@
...
+@inline function is_correct_cell(cell_list::SpatialHashingCellList{<:Any, Nothing},
+                                 coords, cell_index::Array)
+    return coords == cell_index
+end
+
+function spatial_hash(cell::CartesianIndex, list_size)
+    return spatial_hash(Tuple(cell), list_size)
+end
+
+function spatial_hash(cell::NTuple{1, Real}, list_size)
```
efaulhaber, line 90:
```suggestion
# Hash functions according to Ihmsen et al. (2001)
function spatial_hash(cell::NTuple{1, Real}, list_size)
```

src/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,111 @@
...
+
+    return mod(xor(i * 73856093, j * 19349663, k * 83492791), list_size) + 1
+end
+
+function spatial_hash(cell::NTuple{3, Real}, index, list_size)
+    i, j, k = cell
+
+    return mod(xor(i * 73856093, j * 19349663, k * 83492791, index * 7238423947),
+               list_size) + 1
+end
```
efaulhaber, line 111: Is this used anywhere or a relic of my early experiments with linear probing? Can it be removed?

src/nhs_grid.jl
```diff
@@ -441,18 +441,45 @@ function update_grid!(neighborhood_search::Union{GridNeighborhoodSearch{<:Any,
     initialize_grid!(neighborhood_search, y; parallelization_backend)
 end
 
+function check_collision(neighbor_cell_, neighbor_coords, cell_list, nhs)
```
efaulhaber, line 444:
```suggestion
function check_collision(neighbor_cell_, neighbor_coords, cell_list, nhs)
    # This is only relevant for the `SpatialHashingCellList`
```

src/nhs_grid.jl
```diff
@@ -441,18 +441,45 @@ function update_grid!(neighborhood_search::Union{GridNeighborhoodSearch{<:Any,
...
+function check_collision(neighbor_cell_::CartesianIndex, neighbor_coords,
+                         cell_list::SpatialHashingCellList, nhs)
+    (; list_size, collisions, coords) = cell_list
+    neighbor_cell = periodic_cell_index(Tuple(neighbor_cell_), nhs)
+
+    return neighbor_cell != cell_coords(neighbor_coords, nhs)
+end
+
+function check_cell_collision(neighbor_cell_::CartesianIndex,
+                              cell_list, nhs)
```
efaulhaber, line 457:
```suggestion
                              cell_list, nhs)
    # This is only relevant for the `SpatialHashingCellList`
```

src/nhs_grid.jl
```diff
@@ -441,18 +441,45 @@ function update_grid!(neighborhood_search::Union{GridNeighborhoodSearch{<:Any,
...
+    return false
+end
+
+function check_cell_collision(neighbor_cell_::CartesianIndex,
+                              cell_list::SpatialHashingCellList, nhs)
+    (; list_size, collisions, coords) = cell_list
+    neighbor_cell = periodic_cell_index(Tuple(neighbor_cell_), nhs)
+    hash = spatial_hash(neighbor_cell, list_size)
+
+    return collisions[hash] || coords[hash] != neighbor_cell
```
efaulhaber, line 467:
```suggestion
    # `collisions[hash] == true` means points from multiple cells are in this list.
    # `collisions[hash] == false` means points from only one cells are in this list.
    # We could still have a collision though, if this one cell is not `neighbor_cell`,
    # which is possible when `neighbor_cell` is empty.
    return collisions[hash] || coords[hash] != neighbor_cell
```

src/nhs_grid.jl
```diff
@@ -441,18 +441,45 @@ function update_grid!(neighborhood_search::Union{GridNeighborhoodSearch{<:Any,
...
+end
+
 # Specialized version of the function in `neighborhood_search.jl`, which is faster
-# than looping over `eachneighbor`.
+# than looping over eachneighbor`.
```
efaulhaber, line 471:
```suggestion
# than looping over `eachneighbor`.
```

src/nhs_grid.jl
```diff
@@ -441,18 +441,45 @@ function update_grid!(neighborhood_search::Union{GridNeighborhoodSearch{<:Any,
...
-
+                                    neighborhood_search::GridNeighborhoodSearch,
+                                    point, point_coords, search_radius)
+    (; cell_list, periodic_box) = neighborhood_search
     cell = cell_coords(point_coords, neighborhood_search)
 
     for neighbor_cell_ in neighboring_cells(cell, neighborhood_search)
         neighbor_cell = Tuple(neighbor_cell_)
         neighbors = points_in_cell(neighbor_cell, neighborhood_search)
+        cell_collision = check_cell_collision(neighbor_cell_,
```
efaulhaber, line 481:
```suggestion

        # Boolean to indicate if this cell has a collision (only with `SpatialHashingCellList`)
        cell_collision = check_cell_collision(neighbor_cell_,
```

src/nhs_grid.jl
```diff
@@ -471,6 +498,13 @@ end
             if distance2 <= search_radius^2
                 distance = sqrt(distance2)
 
+                # Check if `neighbor_coords` are in the cell `neighbor_cell_`.
+                # For the `SpatialHashingCellList`, this might not be the case
+                # if we have a collision.
```
efaulhaber, line 503:
```suggestion
                # If this cell has a collision, check if this point belongs to this cell
                # (only with `SpatialHashingCellList`).
```

src/nhs_grid.jl
```diff
@@ -441,18 +441,45 @@ function update_grid!(neighborhood_search::Union{GridNeighborhoodSearch{<:Any,
     initialize_grid!(neighborhood_search, y; parallelization_backend)
 end
 
+function check_collision(neighbor_cell_, neighbor_coords, cell_list, nhs)
+    return false
+end
+
+function check_collision(neighbor_cell_::CartesianIndex, neighbor_coords,
```
efaulhaber, line 448:
```suggestion
# Check if `neighbor_coords` belong to `neighbor_cell`, which might not be the case
# with the `SpatialHashingCellList` if this cell has a collision.
function check_collision(neighbor_cell_::CartesianIndex, neighbor_coords,
```

src/nhs_grid.jl
```diff
@@ -441,18 +441,45 @@ function update_grid!(neighborhood_search::Union{GridNeighborhoodSearch{<:Any,
...
+
+    return neighbor_cell != cell_coords(neighbor_coords, nhs)
+end
+
+function check_cell_collision(neighbor_cell_::CartesianIndex,
+                              cell_list, nhs)
+    return false
+end
+
+function check_cell_collision(neighbor_cell_::CartesianIndex,
```
efaulhaber, line 461:
```suggestion
# Check if there is a collision in this cell, meaning there is at least one point
# in this list that doesn't actually belong in this cell.
function check_cell_collision(neighbor_cell_::CartesianIndex,
```

src/nhs_grid.jl
```diff
@@ -471,6 +498,13 @@ end
             if distance2 <= search_radius^2
                 distance = sqrt(distance2)
 
+                # Check if `neighbor_coords` are in the cell `neighbor_cell_`.
+                # For the `SpatialHashingCellList`, this might not be the case
+                # if we have a collision.
+                if check_collision(neighbor_cell_, neighbor_coords, cell_list, neighborhood_search) && cell_collision
```
efaulhaber, line 504:
```suggestion
                if cell_collision && check_collision(neighbor_cell_, neighbor_coords, cell_list, neighborhood_search)
```
This is crucial! This is a short-circuit evaluation, which is evaluated from left to right, meaning you always call the function `check_collision`, and then check the flag `cell_collision`.
With my suggestions, you skip the call to `check_collision` for most cells because `cell_collision` is false.

This should also greatly improve performance.

test/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,75 @@
...
+            neighbors = [Int[] for _ in axes(coords, 2)]
+            foreach_point_neighbor(coords, coords, nhs,
+                                   parallel = false) do point, neighbor,
+                                                        pos_diff, distance
+                push!(neighbors[point], neighbor)
+            end
+
+            @test sort.(neighbors) == neighbors_expected
+        end
+    end
```
efaulhaber, line 50: This is already covered in `test/neighborhood_search.jl`, no?

test/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,75 @@
...
+
+    # Test list behavior with empty cells
+    @testset "Collision Handling with empty cells" begin
+        # The point is in cell (-1, 0) which has a hash collision with cell (-2, -1)
+        point = [-0.05, 0.05]
+        coordinates = hcat(point)
```
efaulhaber, line 56:
```suggestion
        coordinates = [-0.05; 0.05;;]
```

test/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,75 @@
...
+        NDIMS = size(coordinates, 1)
+        n_points = size(coordinates, 2)
+        search_radius = 0.1 + 10 * eps()
+        point_index = 1
+
+        nhs = GridNeighborhoodSearch{2}(; search_radius, n_points,
+                                            cell_list = SpatialHashingCellList{NDIMS}(n_points))
+        initialize_grid!(nhs, coordinates)
+
```
efaulhaber, line 65: Please add a check here that you indeed have a collision. Check that both (-1, 0) and (-2, -1) have the same list.

Also, it would be nice if you had another test with two points in separate cells, which have a collision to cover this case as well.

efaulhaber (changes requested):
Looking good! Most of my requested changes are just comments.

One of my requested changes is performance relevant. Could you please do a quick benchmark after you fixed this against `DictionaryCellList` and `FullGridCellList` and post the results here? Just the time for ~100k particles and for ~10M particles on RAMSES (multithreaded). No need to waste more time on benchmarks at this point, just want to see where we stand now. Performance will become more interesting in the next step when we go to GPUs.

test/cell_lists/spatial_hashing.jl
RubberLanding: Do you want to explicitly compare the two lists or is it enough to check that both have the same hash value?

efaulhaber: Why not both to be sure?
```julia
@test hash1 == hash2
@test cell_list[cell1] == cell_list[cell2]
```
To test both the low-level hash and the cell list `getindex`.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "PointNeighbors"
 uuid = "1c4d5385-0a27-49de-8e2c-43b175c8985c"
 authors = ["Erik Faulhaber <erik.faulhaber@uni-koeln.de>"]
-version = "0.5"
+version = "0.4.9"
```
efaulhaber, line 4:
```suggestion
version = "0.4.9-dev"
```

src/nhs_grid.jl
```diff
@@ -498,15 +510,16 @@ end
...
+                # (only with `SpatialHashingCellList`).
+                if cell_collision &&
+                   check_collision(neighbor_cell_, neighbor_coords, cell_list,
+                                   neighborhood_search)
                     continue
                 end
 
                 # Inline to avoid loss of performance
-                # compared to not using `foreach_point_neighbor`.
+                # compared to not using `foreach_point_neighbor
```
efaulhaber, line 522:
```suggestion
                # compared to not using `foreach_point_neighbor`.
```

test/Project.toml
```diff
@@ -1,12 +0,0 @@
-[deps]
```
efaulhaber, line 1: Please restore this file. This is also the reason why the tests are failing.

test/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,63 @@
...
+        nhs = GridNeighborhoodSearch{2}(; search_radius, n_points,
+                                        cell_list = SpatialHashingCellList{NDIMS}(n_points))
+        initialize_grid!(nhs, coordinates)
+
+        @testset "Test For Collision" begin
+            cell1 = (-1, 0)
+            cell2 = (-2, -1)
+            cell1_hash = PointNeighbors.spatial_hash(cell1, n_points)
+            cell2_hash = PointNeighbors.spatial_hash(cell2, n_points)
```
efaulhaber, line 18:
```suggestion
            cell2_hash = PointNeighbors.spatial_hash(cell2, n_points)
            
            points1 = nhs.cell_list[cell1]
            points2 = nhs.cell_list[cell2]
            @test points1 == points2 == [1]
```

test/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,63 @@
...
```
efaulhaber, line 48:
```suggestion
            cell2_hash = PointNeighbors.spatial_hash(cell2, n_points)
            
            points1 = nhs.cell_list[cell1]
            points2 = nhs.cell_list[cell2]
            @test points1 == points2 == [1, 2]
```

efaulhaber (changes requested):
Just a few minor remarks. Otherwise, this is ready to merge from my side. @LasNikas @svchb please have a look as well.

src/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,108 @@
+"""
+    SpatialHashingCellList{NDIMS}(; list_size)
```
LasNikas, line 2: Can you add this to the table in the "Features" section in the README.md? 
Or is it still considered experimental at this stage?
Additionally, do you have any benchmarks available for the spatial hashing cell list?

efaulhaber: The next step is to make this work with a `DynamicVectorOfVectors` backend. Then it is GPU-compatible and we can run proper benchmarks. Before that, it doesn't really make sense to list it in the README.

svchb: Well in general it might be good to add better docs here and then even do a short JOSS paper to get a pub out of this.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "PointNeighbors"
 uuid = "1c4d5385-0a27-49de-8e2c-43b175c8985c"
 authors = ["Erik Faulhaber <erik.faulhaber@uni-koeln.de>"]
-version = "0.5"
+version = "0.4.9-dev"
```
svchb, line 4: @efaulhaber is this correct?

src/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,108 @@
...
+
+A basic spatial hashing implementation. Similar to [`DictionaryCellList`](@ref), the domain is discretized into cells, 
+and the particles in each cell are stored in a hash map. The hash is computed using the spatial location of each cell 
+[as described by Ihmsen et al. (2001)](@cite Ihmsen2003). By using a hash map, which only stores non-empty cells,
```
svchb, line 6:
```suggestion
[as described by Ihmsen et al. (2001)](@cite Ihmsen2003). By using a hash map, which only stores non-empty cells, 
```
Ihmsen 2001 or 2003?

svchb, line 6:
```suggestion
, as described by Ihmsen et al. (2001)(@cite Ihmsen2003). By using a hash map, which only stores non-empty cells, 
```

src/cell_lists/spatial_hashing.jl
```diff
@@ -0,0 +1,108 @@
+"""
+    SpatialHashingCellList{NDIMS}(; list_size)
+
+A basic spatial hashing implementation. Similar to [`DictionaryCellList`](@ref), the domain is discretized into cells, 
+and the particles in each cell are stored in a hash map. The hash is computed using the spatial location of each cell 
+[as described by Ihmsen et al. (2001)](@cite Ihmsen2003). By using a hash map, which only stores non-empty cells, 
+the domain is effectively infinite. The size of the hash map is recommended to be approximately twice the number of particles to balance memory consumption
```
svchb, line 7:
```suggestion
[as described by Ihmsen et al. (2001)](@cite Ihmsen2003). By using a hash map that stores entries only for non-empty cells, the domain is effectively infinite. The size of the hash map is recommended to be approximately twice the number of particles to balance memory consumption  
```

svchb (changes requested):

src/cell_lists/spatial_hashing.jl
efaulhaber: Not worth a JOSS paper at this point. Once we get GPU support, we can think about it again.

Project.toml
efaulhaber: Yes, we need this for the tests to run because the TP release doesn't support PN 0.5 yet.

src/cell_lists/spatial_hashing.jl
efaulhaber: I think even 2011. @RubberLanding ?

src/cell_lists/spatial_hashing.jl
svchb: yes I meant in the future.

src/cell_lists/spatial_hashing.jl
RubberLanding: I was using [this paper](https://cg.informatik.uni-freiburg.de/publications/2011_CGF_dataStructuresSPH.pdf) which is from 2010.

svchb: yes for references only the official publication date counts. If you check the bibtex:
@inproceedings{ihmsen2011parallel,
  title={A parallel SPH implementation on multi-core CPUs},
  author={Ihmsen, Markus and Akinci, Nadir and Becker, Markus and Teschner, Matthias},
  booktitle={Computer Graphics Forum},
  volume={30},
  number={1},
  pages={99--112},
  year={2011},
  organization={Wiley Online Library}
}

efaulhaber (approved):
LGTM, but the format check is failing.

svchb:
@efaulhaber If you are starting to add references you should also add a bib file and setup stub docs.

svchb:
@RubberLanding For the formatting test to not fail you need to use JuliaFormatter.jl and apply it e.g.

```
using JuliaFormatter
format(".")
```

efaulhaber (approved):

svchb (approved):

PR #105, opened 2025-04-10
Add function `requires_update`

efaulhaber:


efaulhaber:
@LasNikas I also added `requires_resizing`, which you can use with the buffer in TrixiParticles:
```julia
if requires_resizing(nhs)
    # Use `Inf` values for inactive particles
else
    update!(nhs, particles, active_neighbors)
end
```

svchb (approved):

LasNikas (approved):

PR #109, opened 2025-05-05
Add full support for parallelization backends

efaulhaber:
- Added the `SerialBackend`.
- Changed `@threaded` API to only allow backends as first argument and not arrays. This is a breaking change.
- Added the kwarg `parallelization_backend` to `initialize!` and to docs (and removed experimental warning).
- Renamed the kwarg `parellel` of `foreach_point_neighbor` to `parallelization_backend` and removed support for booleans like `parallel = true`. This is a breaking change.
- Added better error message for `FullGridCellList` out of bounds (on CPUs only).
- Removed `requires_resizing` (@LasNikas requested it but didn't need it at the end). This is a breaking change.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PointNeighbors"
 uuid = "1c4d5385-0a27-49de-8e2c-43b175c8985c"
 authors = ["Erik Faulhaber <erik.faulhaber@uni-koeln.de>"]
-version = "0.5"
+version = "0.4.9-dev"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/neighborhood_search.jl
+++ b/src/neighborhood_search.jl
@@ -55,8 +55,8 @@ in this case to avoid unnecessary updates.
 The first flag in `points_moving` indicates if points in `x` are moving.
 The second flag indicates if points in `y` are moving.
 
-If the neighborhood search update supports parallelization, the keyword argument
-`parallelization_backend` can be used to specify the parallelization backend.
+If the neighborhood search type supports parallelization, the keyword argument
+`parallelization_backend` can be used to specify a parallelization backend.
 See [`@threaded`](@ref) for a list of available backends.
 
 See also [`initialize!`](@ref).
--- a/src/neighborhood_search.jl
+++ b/src/neighborhood_search.jl
@@ -23,8 +23,8 @@ all points in `y` whose distances to that point are smaller than the search radi
 `x` and `y` are expected to be matrices, where the `i`-th column contains the coordinates
 of point `i`. Note that `x` and `y` can be identical.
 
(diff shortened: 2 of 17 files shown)


efaulhaber:
Tests are failing until https://github.com/trixi-framework/TrixiParticles.jl/pull/790 is merged, which is based on this PR, so we will need to ignore these fails for now.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "PointNeighbors"
 uuid = "1c4d5385-0a27-49de-8e2c-43b175c8985c"
 authors = ["Erik Faulhaber <erik.faulhaber@uni-koeln.de>"]
-version = "0.5"
+version = "0.4.9-dev"
```
svchb, line 4:
```suggestion
version = "0.5.1-dev"
```

src/neighborhood_search.jl
```diff
@@ -32,19 +23,27 @@ all points in `y` whose distances to that point are smaller than the search radi
 `x` and `y` are expected to be matrices, where the `i`-th column contains the coordinates
 of point `i`. Note that `x` and `y` can be identical.
 
+If the neighborhood search initialization supports parallelization, the keyword argument
+`parallelization_backend` can be used to specify the parallelization backend.
+See [`@threaded`](@ref) for a list of available backends.
```
svchb, line 28:
```suggestion
If the neighborhood search type supports parallelization, the keyword argument
`parallelization_backend` can be used to specify a parallelization backend.
See [`@threaded`](@ref) for a list of available backends.
```

src/neighborhood_search.jl
```diff
@@ -56,20 +55,15 @@ in this case to avoid unnecessary updates.
...
-    multithreading backend. This feature is currently considered experimental!
-
-    Possible parallelization backends are:
-    - [`ThreadsDynamicBackend`](@ref) to use `Threads.@threads :dynamic`
-    - [`ThreadsStaticBackend`](@ref) to use `Threads.@threads :static`
-    - [`PolyesterBackend`](@ref) to use `Polyester.@batch`
-    - `KernelAbstractions.Backend` to launch a GPU kernel
+If the neighborhood search update supports parallelization, the keyword argument
+`parallelization_backend` can be used to specify the parallelization backend.
+See [`@threaded`](@ref) for a list of available backends.
```
svchb, line 60:
```suggestion
If the neighborhood search type supports parallelization, the keyword argument
`parallelization_backend` can be used to specify a parallelization backend.
See [`@threaded`](@ref) for a list of available backends.
```

svchb (changes requested):

Project.toml
efaulhaber: I need to set 0.4.X to make the tests run, as the current TrixiP release is not compatible with PN 0.5.

svchb (approved):

LasNikas (approved):

PR #115, opened 2025-05-14
Add option to skip inactive points in `foreach_point_neighbor` and `update` without assembling an iterator

efaulhaber:


src/neighborhood_search.jl
```diff
@@ -181,14 +182,20 @@ function foreach_point_neighbor(f::T, system_coords, neighbor_coords, neighborho
...
+            @inbounds foreach_neighbor(f, system_coords, neighbor_coords,
+                                    neighborhood_search, point)
+        end
     end
 
     return nothing
 end
 
+@inline point_active(::Nothing, _) = true
+@inline point_active(points_active, point) = points_active[point]
```
LasNikas, line 197:
```suggestion
@inline point_active(points_active, point) = points_active[point] == true
```
In case we use UInt32 instead of Bool. (CUDA doesn't support atomic operations on bool)

src/nhs_grid.jl
```diff
@@ -201,7 +201,7 @@ end
 
 function initialize_grid!(neighborhood_search::GridNeighborhoodSearch, y::AbstractMatrix;
                           parallelization_backend = default_backend(y),
-                          eachindex_y = axes(y, 2))
+                          eachindex_y = axes(y, 2), points_active = nothing)
```
LasNikas, line 204: key word is missing in the function above
```julia
function initialize!(neighborhood_search::GridNeighborhoodSearch,
                     x::AbstractMatrix, y::AbstractMatrix;
                     parallelization_backend = default_backend(x),
                     eachindex_y = axes(y, 2), points_active = nothing)
    initialize_grid!(neighborhood_search, y; parallelization_backend, eachindex_y)
end
```

PR #147, opened 2026-03-20
Use unsafe_indices

vchuravy:
Based on AcceleratedKernels.jl


sloede:
We like "unsafe" if it implies raw speed 😈

efaulhaber:
@vchuravy can you explain why? I benchmarked this and found absolutely no difference in performance.

PR #158, opened 2026-05-25
Add functions `mapreduce_neighbor` and `mapreduce_neighbor_unsafe`

efaulhaber:
This PR allows us to convert the awkward
```julia
dv_particle = Ref(zero(v_a))

foreach_neighbor(...) do particle, neighbor, pos_diff, distance
    ...
    dv_particle[] += result
end
```
to a cleaner
```julia
dv_particle = mapreduce_neighbor(+, ..., init=zero(v_a)) do particle, neighbor, pos_diff, distance
    ...
    return result
end
```

in https://github.com/trixi-framework/TrixiParticles.jl/pull/1211.

This is necessary for vectorization on the CPU.

### DIFF
--- a/src/neighborhood_search.jl
+++ b/src/neighborhood_search.jl
@@ -330,7 +330,10 @@ end
 Apply `f(i, j, pos_diff, d)` to every neighbor of `point` and reduce the results with
 the binary operator `op`, analogous to `mapreduce(f, op, collection)`.
 
-The keyword argument `init` is required and is returned if `point` has no neighbors.
+The keyword argument `init` is required. It provides the starting value for
+the reduction. Choose `init` so that combining it with the first mapped neighbor value
+using `op` gives the desired result, typically the identity element of `op`.
+
 This method performs the same bounds checks as [`foreach_neighbor`](@ref).
 See [`mapreduce_neighbor_unsafe`](@ref) for a version that skips all bounds checks.
 """
--- a/src/neighborhood_search.jl
+++ b/src/neighborhood_search.jl
@@ -262,13 +262,17 @@ See [`foreach_neighbor_unsafe`](@ref) for a version that skips all bounds checks
                      point, point_coords, search_radius)
 end
 
+@inline foreach_neighbor_op(::Any, ::Any) = nothing
+
 # This is a function barrier to prevent the `@inbounds` in `foreach_neighbor`
 # from propagating into the neighbor loop, which is not safe.
 @inline function foreach_neighbor(f, neighbor_coords,
                                   neighborhood_search::AbstractNeighborhoodSearch,
                                   point, point_coords, search_radius)
-    foreach_neighbor_inner(f, neighbor_coords, neighborhood_search,
-                           point, point_coords, search_radius)
+    mapreduce_neighbor_inner(f, foreach_neighbor_op,
+                             neighbor_coords, neighborhood_search,
+                             point, point_coords, search_radius, nothing)
+    return nothing
 end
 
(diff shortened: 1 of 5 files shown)


src/neighborhood_search.jl
```diff
@@ -312,20 +316,82 @@ Note that all these bounds checks are safe to skip if
...
+    mapreduce_neighbor(f, op, system_coords, neighbor_coords,
+                       neighborhood_search::AbstractNeighborhoodSearch, point;
+                       init, search_radius = search_radius(neighborhood_search))
+
+Apply `f(i, j, pos_diff, d)` to every neighbor of `point` and reduce the results with
+the binary operator `op`, analogous to `mapreduce(f, op, collection)`.
+
+The keyword argument `init` is required and is returned if `point` has no neighbors.
+This method performs the same bounds checks as [`foreach_neighbor`](@ref).
+See [`mapreduce_neighbor_unsafe`](@ref) for a version that skips all bounds checks.
```
svchb:
```suggestion
The keyword argument `init` is required. It provides the starting value for
the reduction and is returned unchanged if `point` has no neighbors. Choose
`init` so that combining it with the first mapped neighbor value using `op`
gives the desired result, typically the identity element of `op`.
```

test/neighborhood_search.jl
```diff
@@ -315,12 +324,69 @@
...
+                        mapreduce_neighbor(+, coords, coords, nhs, point;
+                                           init = 0) do point_, neighbor,
+                                                        pos_diff, distance
+                            point_ == point || error("incorrect point index")
+                            neighbor
+                        end
+                    end
+
+                    @test neighbor_sums == sum.(neighbors_expected)
+
```
svchb, line 400:
```suggestion
@test mapreduce_neighbor(+, coords, coords, nhs, point;
                         init = 12345,
                         search_radius = 0) do point, neighbor, pos_diff, distance
    neighbor
end == 12345
```

svchb (changes requested):

src/neighborhood_search.jl
efaulhaber: `point` always has neighbors: at least itself.

test/neighborhood_search.jl
```diff
@@ -370,12 +379,79 @@
...
+                    # Test that `init` is used as starting value for the reduction.
+                    point = 1
+                    empty_sum = mapreduce_neighbor_unsafe(+, coords, coords, nhs, point;
+                                                          init = 123) do point_, neighbor,
+                                                                         pos_diff, distance
+                        return 0
+                    end
+
+                    @test empty_sum == 123
```
svchb:
```suggestion
                   # Test the reduction over an empty neighborhood.
                    # Using a non-neutral `init` here is intentional: for an empty
                    # neighborhood, `init` must be returned unchanged.
                    empty_nhs = copy_neighborhood_search(nhs, search_radius,
                                                         size(coords, 2))
                    initialize!(empty_nhs, coords, coords; eachindex_y = Int[])

                    point = first(axes(coords, 2))

                    result = mapreduce_neighbor_unsafe(
                        (point, neighbor, pos_diff, distance) ->
                            error("mapper must not be called for an empty neighborhood"),
                        (a, b) ->
                            error("reduction operator must not be called for an empty neighborhood"),
                        coords, coords, empty_nhs, point;
                        init = 123
                    )

                   @test result == 123
```

svchb (changes requested):

svchb (approved):

PR #171, opened 2026-08-25
Add CellListMap.jl neighborhood search extension

lmiq:
Disclaimer: This PR used Claude Code extensively. 

It might not be intended for merging, if not of interest for PointNeighbors.jl. I was mostly interested in knowing if the performances justified migrating my code to a different backend. 

## Summary

Wraps CellListMap.jl's ParticleSystem/pairwise! API as a package extension providing CellListMapNeighborhoodSearch, superseding the stale approach in trixi-framework#8 (whose merge base predates both the parallelization-backend refactor here and CellListMap.jl's own API redesign since then).

Includes a correctness test suite against TrivialNeighborhoodSearch and a benchmark at atomic water density with a 12 Angstrom cutoff.

Supersedes the stale approach in #8. That branch's merge base is ~100 commits behind
`main` and predates two rounds of API drift on both sides:
- PointNeighbors' `foreach_point_neighbor` moved from a `parallel::Bool` kwarg to the
  `parallelization_backend::ParallelizationBackend` interface used everywhere else now.
- CellListMap.jl's own public API moved from the low-level `Box`/`CellList`/`CellListPair`/
  `map_pairwise!` calls to the higher-level `ParticleSystem`/`pairwise!`/`update!` API, which
  is actually a better fit here: non-periodic bounding-box growth, cell list rebuilding, and
  batch-count updates are all handled internally, so this wrapper needs no manual box/cell
  bookkeeping.

This is a fresh implementation against current `main`, not a rebase, reusing the wrapper
shape and the `points_equal_neighbors` single-set optimization from #8's discussion.

## Changes

- `ext/PointNeighborsCellListMapExt.jl`: new `CellListMapNeighborhoodSearch`, implementing
  the full `AbstractNeighborhoodSearch` interface (`search_radius`, `initialize!`, `update!`,
  `requires_update`, `foreach_point_neighbor`/`foreach_point_neighbor_unsafe`,
  `copy_neighborhood_search`) on top of `CellListMap.ParticleSystem`.
- `test/cell_list_map.jl`: correctness tests against `TrivialNeighborhoodSearch` (2D/3D, both
  `points_equal_neighbors` modes, two distinct point sets, `initialize!`+`update!` including a
  large-move case that forces the non-periodic bounding box to grow, template/copy usage, and
  a `distance == search_radius` boundary case).
- `benchmarks/cell_list_map_water.jl`: benchmark at atomic water density (0.1003 Å⁻³) and a
  12 Å cutoff (~726 neighbors/particle) — much denser than PointNeighbors' usual SPH
  benchmarks, and the regime where CellListMap.jl should have an edge per the discussion on #8.

## Benchmark results

At 110,592 particles (water density, 12 Å cutoff):

| | count_neighbors | n_body |
|---|---|---|
| `GridNeighborhoodSearch` (`DictionaryCellList`) | 73.6 ms | 101.9 ms |
| `GridNeighborhoodSearch` (`FullGridCellList`) | 60.4 ms | 96.0 ms |
| `CellListMapNeighborhoodSearch` | **39.9 ms** | **73.7 ms** |

CellListMap.jl is fastest at every size tested, confirming that it wins in this
denser, MD-typical regime even though #8 found it losing on sparser SPH benchmarks.


lmiq:
Found and fixed a real bug while writing a minimal repro for the benchmark: at water
density with a 12 Å cutoff, average cell occupancy (~173) exceeds `FullGridCellList`'s
default `max_points_per_cell = 100`. The bounds check that would normally catch this
(`"cell list is full. Use a larger max_points_per_cell."`) is skipped in the parallel
update path for performance, so points beyond the limit were **silently dropped**
instead of raising an error.

I caught it because `FullGridCellList`'s neighbor count didn't match
`DictionaryCellList`/`CellListMap`'s on a manual repro, and confirmed it with
`--check-bounds=yes`, which does throw. This meant `benchmarks/cell_list_map_water.jl`
had `FullGridCellList` silently doing less work than the other two, making it look
more competitive than it actually is.

Fixed in the latest commit by raising `max_points_per_cell` with a safety margin.
Corrected numbers at 110,592 particles (water density, 12 Å cutoff):

| | count_neighbors | n_body |
|---|---|---|
| `GridNeighborhoodSearch` (`DictionaryCellList`) | 67.2 ms | 97.8 ms |
| `GridNeighborhoodSearch` (`FullGridCellList`) | 68.2 ms | 96.7 ms |
| `CellListMapNeighborhoodSearch` | **41.0 ms** | **75.0 ms** |

Conclusion is unchanged (CellListMap.jl fastest in this denser, MD-typical regime), but
`FullGridCellList` is now doing the correct amount of work rather than an artificially
reduced one.

lmiq:
For the records, this is a MWE comparison:

```julia
using PointNeighbors
using CellListMap: CellListMap
using BenchmarkTools
using Random

# Atomic number density of water (O + 2H per molecule), in Å⁻³
density = 0.1003
cutoff = 12.0

n_points = 50_000
domain_size = (n_points / density)^(1 / 3)

Random.seed!(1)
coords = domain_size .* rand(3, n_points)

min_corner = zeros(3)
max_corner = fill(domain_size, 3)

# Average occupancy of a `search_radius`-sized cell is `density * cutoff^3` (≈173 here);
# FullGridCellList's default `max_points_per_cell = 100` silently drops points beyond
# that limit (only caught with `--check-bounds=yes`), so it must be raised for this density.
max_points_per_cell = round(Int, 4 * density * cutoff^3)

neighborhood_searches = [
    GridNeighborhoodSearch{3}(; search_radius = cutoff, n_points),
    GridNeighborhoodSearch{3}(; search_radius = cutoff, n_points,
                              cell_list = FullGridCellList(; min_corner, max_corner,
                                                           search_radius = cutoff,
                                                           max_points_per_cell)),
    CellListMapNeighborhoodSearch(3; search_radius = cutoff)
]

names = ["GridNeighborhoodSearch (DictionaryCellList)",
         "GridNeighborhoodSearch (FullGridCellList)",
         "CellListMapNeighborhoodSearch"]

function count_neighbors!(n_neighbors, coords, nhs)
    n_neighbors .= 0
    foreach_point_neighbor(coords, coords, nhs) do i, j, pos_diff, distance
        n_neighbors[i] += 1
    end
end

for (name, nhs) in zip(names, neighborhood_searches)
    initialize!(nhs, coords, coords)

    n_neighbors = zeros(Int, n_points)
    time = @belapsed count_neighbors!($n_neighbors, $coords, $nhs)

    println(rpad(name, 45), " ", BenchmarkTools.prettytime(time * 1e9),
            " (", round(sum(n_neighbors) / n_points, digits = 1), " neighbors/particle)")
end
```

efaulhaber:
> Found and fixed a real bug while writing a minimal repro for the benchmark: at water
density with a 12 Å cutoff, average cell occupancy (~173) exceeds FullGridCellList's
default max_points_per_cell = 100. The bounds check that would normally catch this
("cell list is full. Use a larger max_points_per_cell.") is skipped in the parallel
update path for performance, so points beyond the limit were silently dropped
instead of raising an error.

Thanks for reporting!

Regarding the performance, I basically only focused on GPU performance in the past years, and only took the extra CPU performance that came for free with optimizing the GPU. I will probably work on the CPU performance in the next weeks. Together with a colleague, we optimized and vectorized a MWE for the SPH neighbor loop, and got a (serial) 9.6x speedup on Intel Xeon, 3.7x on Apple M2 Pro, and 12x on AMD Epyc.

Do I understand your benchmarks correctly? It looks like FullGridCellList is slightly faster for the n-body benchmark with a typical SPH search radius, but in your case with a much larger search radius, CellListMap is significantly faster. Is that correct?

lmiq:
I'm not sure. Apparently more or less the same trends are observed for SPH-densities (see below).

I think this is more or less expected: the build and update phases (particularly updating) is much better with PointNeighbors, in CellListMap we have to rebuild everything from scratch. The mapping phase is fast in CellListMap because it is vectorized to some extent (we do have a sorting mechanism of the interactions to be computed between particles of vicinal cells, but if vectorization occurs that is dependent on the Julia compiler). So I expect that if you manage to vectorize your calculations as well, the performance gap in that phase will be reduced. 

At the same time, I think (I have not tested that) the PointNeighbors approach is much more GPU friendly and should scale much better for much larger number of CPUs, which is very likely a good tradeoff for large scale SPH simulations. 

The machine used here is:  CPU: Intel Core i9-14900KF / 14 threads


**Native comparison at SPH-typical density (`search_radius_factor = 3.0`, ~27 particles/cell)**

Same setup as the water-density comparison above (each library used natively, `PolyesterBackend()`/`parallel=true`, 14 threads, Σ 1/d over all pairs for the mapping phase), rerun at PointNeighbors.jl's own "typical SPH" density instead of atomic water density, for direct comparison. All three implementations again agree on the computed sum to float-summation-order precision at every size.

### 1. Build

| particles | DictionaryCellList | FullGridCellList | CellListMap |
|---|---|---|---|
| 1,728 | 30.4 μs | 14.8 μs | 49.5 μs |
| 6,859 | 140.9 μs | 35.8 μs | 113.2 μs |
| 27,000 | 561.3 μs | 109.7 μs | 456.5 μs |
| 110,592 | 2.688 ms | 363.2 μs | 2.018 ms |

Same pattern as water density: CellListMap sits between the two Grid variants, clearly behind `FullGridCellList`.

### 2. Update

| particles | DictionaryCellList | FullGridCellList | CellListMap |
|---|---|---|---|
| 1,728 | 3.1 μs | 14.4 μs | 38.9 μs |
| 6,859 | 10.2 μs | 35.5 μs | 83.8 μs |
| 27,000 | 32.8 μs | 109.9 μs | 289.3 μs |
| 110,592 | 148.9 μs | 363.3 μs | **1.064 ms** |

Same story as water density, more pronounced: CellListMap is slowest at every size, and the gap to `DictionaryCellList` widens to **7.1x** at 110,592 particles (vs. 4.5x at water density). Consistent with more, smaller cells at this lower density adding more per-particle bookkeeping overhead to the unconditional full-rebuild `update!` does regardless of how many particles actually moved.

### 3. Map (Σ 1/d over all pairs — native reduction on both sides)

| particles | DictionaryCellList | FullGridCellList | CellListMap |
|---|---|---|---|
| 1,728 | 218.6 μs | 185.9 μs | 93.4 μs |
| 6,859 | 1.319 ms | 1.082 ms | **391.5 μs** |
| 27,000 | 5.944 ms | 5.003 ms | **1.463 ms** |
| 110,592 | 29.661 ms | 22.640 ms | **6.422 ms** |

CellListMap's mapping advantage holds here too, and by a similar or larger margin than at water density: 2.3x at the smallest size, growing to 3.5-4.6x at the largest (vs. 1.3x-4.9x at water density). The native symmetry exploitation and parallel-scaling advantages described in the water-density comparison aren't specific to that denser regime.

**Bottom line:** density changes the absolute numbers, not the shape of the story. Build is a wash-to-behind-`FullGridCellList`; update is CellListMap's clear and *worsening* weak point at lower density; mapping is a clear, density-independent win for CellListMap when used natively for a reduction-shaped computation.

